PR #20, opened 2021-09-08
Add per-segment regressors length check

julia-shenshina:


etna/datasets/tsdataset.py
```diff
@@ -129,13 +130,30 @@ def make_future(self, future_steps: int) -> "TSDataset":
...
+            target = df[segment]["target"].dropna()
+            exog_regressor_columns = [x for x in set(df_exog[segment].columns) if x.startswith("regressor")]
+            for series in exog_regressor_columns:
+                exog_series = df_exog[segment][series].dropna()
+                if target.index.min() < exog_series.index.min():
+                    raise ValueError(
+                        f"All the regressor series should start not later than corresponding 'target'."
+                        f"Series {series} of segment {segment} have not enough history: "
+                        f"{target.index.min()} < {exog_series.index.min()}."
+                    )
```
iKintosh, line 147: This way we won't show all problematic segments. It is not huge problem right now, but for convenience I'd like to see all segments contain this problem

iKintosh (approved):

Mr-Geekman (approved):

PR #27, opened 2021-09-11
WIP: Remove black; Add Mypy and flake8 plugins

iKintosh:
Black is great tool for python code formatting, however it doesn't follow all the PEP rules, so I suggest we don't use it.
Instead we use some of flake8 plugins.
And to solve problem with us forgetting to set up types correctly I suggest we use Mypy.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 [flake8]
-ignore = F, E203, W605, E501, W503, D100, D104
+ignore = F, E203, W605, E501, W503, D100, D104, C812, SIM106, SIM113, SIM119
 max-line-length = 121
 max-complexity = 18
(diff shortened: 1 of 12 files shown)


iKintosh:
Added 
```
 flake8-commas = {version = "^2.0.0", optional = true}
 flake8-comprehensions = {version = "^3.6.1", optional = true}
 flake8-bugbear = {version = "^21.4.3", optional = true}
 flake8-eradicate = {version = "^1.1.0", optional = true}
 flake8-broken-line = {version = "^0.3.0", optional = true}
 flake8-return = {version = "^1.1.3", optional = true}
 flake8-simplify = {version = "^0.14.1", optional = true}
 flake8-class-attributes-order = {version = "^0.1.2", optional = true}
 mypy = {version = "^0.910", optional = true}
```

PR #310, opened 2021-11-24
MRMR hand-made

alex-hse-repository:

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves an existing feature)
- [x] New feature (non-breaking change which adds functionality)


etna/analysis/feature_selection/mrmr.py
```diff
@@ -0,0 +1,59 @@
...
+        num of regressors to select; if there are not enough regressors, then all will be selected
+
+    Returns:
+    -------
+    selected_features: List[str]
+        list of `top_k` selected regressors, sorted by their importance
+    """
+    x = x.dropna(axis=1, how="all")
+    relevance_table = x.apply(lambda col: f_classif(col[~col.isna()].to_frame(), y[~col.isna()])[0][0])
+    relevance_table = relevance_table[relevance_table > 0]
```
julia-shenshina, line 31: why do you need this filter?

etna/analysis/feature_selection/mrmr.py
```diff
@@ -0,0 +1,59 @@
...
+    redundancy_table = pd.DataFrame(FLOOR, index=all_features, columns=all_features)
+    top_k = min(top_k, len(all_features))
+
+    for i in range(top_k):
+        score_numerator = relevance_table.loc[not_selected_features]
+        score_denominator = pd.Series(1, index=not_selected_features)
+        if i > 0:
+            last_selected_feature = selected_features[-1]
+            redundancy_table.loc[not_selected_features, last_selected_feature] = (
+                x[not_selected_features].corrwith(x[last_selected_feature]).abs().clip(FLOOR).fillna(FLOOR)
```
julia-shenshina, line 46: why there can be Nans?

julia-shenshina, line 46: why do you need to clip here?

etna/analysis/feature_selection/mrmr.py
```diff
@@ -0,0 +1,59 @@
...
+            )
+            score_denominator = (
+                redundancy_table.loc[not_selected_features, selected_features]
+                .mean(axis=1)
+                .round(5)
```
julia-shenshina, line 51: why do you need to round it and fill with `inf`?

etna/analysis/feature_selection/mrmr.py
```diff
@@ -0,0 +1,59 @@
...
+                .replace(1.0, float("Inf"))
+            )
+        score = score_numerator / score_denominator
```
julia-shenshina, line 54: can be denominator equal to zero? dont we need smth like `nominator / (denominator + eps)` here?

etna/transforms/feature_importance.py
```diff
@@ -94,3 +100,108 @@ def fit(self, df: pd.DataFrame) -> "TreeFeatureSelectionTransform":
         weights = self._get_regressors_weights(df)
         self.selected_regressors = self._select_top_k_regressors(weights, self.top_k)
         return self
+
+
+class MRMRFeatureSelectionTransform(Transform):
```
julia-shenshina, line 105: we have a base class https://github.com/tinkoff-ai/etna/blob/master/etna/transforms/feature_selection.py#L9 now that implements `_get_regressor` and `transform` methods

etna/analysis/feature_selection/mrmr.py
Ama16: It count relevance for all regressors separately and keep only regressors that have positive relevance

etna/analysis/feature_selection/mrmr.py
Ama16: because where was clip before, all values are >= eps

etna/analysis/feature_selection/mrmr.py
Ama16: see below

etna/analysis/feature_selection/mrmr.py
Ama16: because if mean of correlation is 1 it means this regressor is redundant and we dont need use it. About roud: idk but maybe to avoid minor inaccuracies

etna/analysis/feature_selection/mrmr.py
julia-shenshina: 🤔 
if we need it just to prevent division by zero, than `eps` in division looks like the better way because it can be set in args like [here](https://github.com/tinkoff-ai/etna/blob/master/etna/metrics/functional_metrics.py#L9) and doesn't confuses the denominator computation

etna/analysis/feature_selection/mrmr.py
julia-shenshina: is it just about optimization of number of series we work with?

alex-hse-repository: This is how it was implemented in mrmr library)

etna/analysis/feature_selection/mrmr.py
alex-hse-repository: This is how it was implemented in mrmr library)

PR #820, opened 2022-07-27
Add in_column attribute to all the transforms

### LINKED ISSUE #810: Add `in_column` attribute to all the transforms
### 🚀 Feature Request

Add `in_column` attribute to all the transforms, if there is no such attribute yet. This attribute will be used by the transform to get only the necessary columns from the dataset.

### Proposal

1. Inspect the current transforms and make sure that all of them has attribute `in_column: Union[str, List[str], Literal["all"]]`
2. For transforms from timestamp folder create dummy in_column with default value "target"
3. For `FilterFeaturesTransform` create appropriate in_column base on include/exclude
- If include is not None in_column="all"
- If exclude is not None use in_column=exclude + "target"
5. For feature selection transform create in_column based on features_to_use attribute
6. For segment encoders create in_column with value "target"
7. For nn set in_columns to literal "all"

For the transforms with the constant in_column make in the class attribute, do not initialize it in the constructor

### Test cases

Create test, then check that `in_column` value is generated correctly for transforms from feature_selection folder

### Additional context

_No response_

alex-hse-repository:

## Closing issues

closes #810

### DIFF
--- a/etna/transforms/feature_selection/filter.py
+++ b/etna/transforms/feature_selection/filter.py
@@ -3,11 +3,14 @@ from typing import Sequence
 
 import pandas as pd
 
+from etna.transforms.base import DymmyInColumnMixin
 from etna.transforms.base import Transform
 
 
-class FilterFeaturesTransform(Transform):
+class FilterFeaturesTransform(Transform, DymmyInColumnMixin):
     """Filters features in each segment of the dataframe."""
+
+    in_column = "all"
 
     def __init__(
         self,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,30 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## TSDataset 2.0
+### Added
+-
+- 
+- 
+- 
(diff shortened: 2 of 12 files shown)


etna/transforms/feature_selection/filter.py
```diff
@@ -3,12 +3,15 @@
 
 import pandas as pd
 
+from etna.transforms.base import DymmyInColumnMixin
 from etna.transforms.base import Transform
 
 
-class FilterFeaturesTransform(Transform):
+class FilterFeaturesTransform(Transform, DymmyInColumnMixin):
```
martins0n, line 10: Do we really need DymmyInColumnMixin?

martins0n (changes requested):
* You wanted to make tests for transforms for `feature_selection`?
* May be we dont need this Mixin. It seems we can add dummy `in_column` in base `Transform` for example or add `in_column` attribute separately for all classes. We don't need extra class for it.

etna/transforms/feature_selection/filter.py
alex-hse-repository: This mixin indicates the classes with fake or constant in_column, I guess we can leave it now and remove in the future if we do't need it

alex-hse-repository: I though there would be some additional logic in in_column generation for feature_selection, however there isn't

martins0n (approved):

PR #827, opened 2022-07-29
DeadlineMovingAverageModel 

### LINKED ISSUE #801: Add DeadlineMovingAverageModel
### 🚀 Feature Request

Sometimes naive models are not good enough because there are different numbers of days in month/year. It means that we can't use naive models to predict events that happen every 20th day of the month. 

### Add Model
Add models to handle these cases:
```
class DeadlineMovingAverageModel(PerSegmentModel):
 def __init__(window: int = 3, seasonality: str = "month"):
    ...
```
- valid values for `seasonality` are `month`, `year`; raise ValueError if given another one
- add Example section with model usage example
```
  def fit(ts: TSDataset) -> "DeadlineMovingAverageModel": ...
```
- model works only with TSDataset with hourly/daily frequency; raises ValueError otherwise
- model checks that all the given series contain enough points (at least `window` number of `seasonality` intervals)
- at `fit` step model saves last `window` of `seasonality` points for each segment
```
  def forecast(ts: TSDataset) -> TSDataset: ...
```
- at `forecast` step model builds the forecast: for each timestamp of forecast compute average value of `window` points with the same day of month (or the same hour of the same month day in case of hour freq in dataset) from saved history

### corner case: 
  - problem: we build forecast for the May 31st; there is no corresponding timestamp in previous month because there are only 30 days in April
  - solution: use April 30th for prediction

### Test cases

- add test cases to check the main logic of the model
- add test cases to check corner case

### Additional context

_No response_

DBcreator:

## Closing issues

Closes #801


etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+                raise ValueError(
+                    "Given series is too short for chosen shift value. Try lower shift value, or give" "longer series."
+                )
+        else:
+            first_index = self.dates.iloc[-1] - pd.DateOffset(years=self.window)
+            if first_index < self.dates.iloc[0]:
+                raise ValueError(
+                    "Given series is too short for chosen shift value. Try lower shift value, or give" "longer series."
+                )
+
```
julia-shenshina, line 71: looks like we can do smth like
```
if self.seasonality == "month":
    first_index = self.dates.iloc[-1] - pd.DateOffset(months=self.window)
else:
    first_index = self.dates.iloc[-1] - pd.DateOffset(years=self.window)
if first_index < self.dates.iloc[0]:
    raise ValueError(
        "Given series is too short for chosen shift value. Try lower shift value, or give" "longer series."
    )
```
to avoid duplicated code

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        self.series = targets.loc[self.dates >= first_index]
+        self.shift = len(self.series)
+
+        # ???
+        if targets.name is not None:
```
julia-shenshina, line 76: why do we need this `if`?

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+            self.name = targets.name
+        return self
+
+    def predict(self, df: pd.DataFrame):
```
julia-shenshina, line 80:
```suggestion
    def predict(self, df: pd.DataFrame) -> np.ndarray:
```

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        ----------
+        df: pd.DataFrame
+            Used only for getting the horizon of forecast and timestamps
+
+        Returns
+        -------
+        :
+            Array with predictions.
+        """
+        timestamp = df["timestamp"]
```
julia-shenshina, line 94: it is a column of timestamps so it should be name `timestamps`

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        res = np.zeros((len(df), 1))
+        for w in range(1, self.window + 1):
+            prev_dates = []
```
julia-shenshina, line 97: do you need to declare `prev_dates` this way here?
in this row it is a list but then it becomes pd.Series https://github.com/tinkoff-ai/etna/pull/827/files#diff-2c2172cf59bd50407e53f6aa48b771cc84df5e644dfa9a464156f823d91930d9R99

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        Initialize seasonal moving average model.
+
+        Length of remembered tail of series is ``window * seasonality``.
+
+        Parameters
+        ----------
+        window: int
+            Number of values taken for forecast for each point.
+        seasonality: int
+            Lag between values taken for forecast.
```
julia-shenshina, line 129: add available seasonalities to the docs please

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+        return res
+
+
+class DeadlineMovingAverageModel(PerSegmentModel):
+    """Seasonal moving average model that uses exact previous dates to predict."""
+
+    def __init__(self, window: int = 3, seasonality: str = "month"):
+        """
+        Initialize seasonal moving average model.
```
julia-shenshina, line 120: smth went wrong with docs here

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        seasonality: int
+            Lag between values taken for forecast.
+        """
+        self.window = window
+        self.seasonality = seasonality
+        super(DeadlineMovingAverageModel, self).__init__(
+            base_model=_DeadlineMovingAverageModel(window=window, seasonality=seasonality)
+        )
+
+    def get_model(self) -> Dict[str, "DeadlineMovingAverageModel"]:
```
julia-shenshina, line 137: could you please add a test to check this method works?

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+            prev_dates = []
+            if self.seasonality == "month":
+                prev_dates = timestamp - pd.DateOffset(months=w)
+            else:
+                prev_dates = timestamp - pd.DateOffset(years=w)
+            values = []
+            for date in prev_dates:
+                value = self.series.loc[self.dates == date].values
+                values.append(value)
+            res += np.array(values)
```
julia-shenshina, line 106: shouldn't it be smth like `res += np.array(values).sum()`?

tests/test_models/test_deadline_ma.py
```diff
@@ -0,0 +1,108 @@
...
+    df = pd.concat([df1, df2]).reset_index(drop=True)
+    df = TSDataset.to_dataset(df)
+    tsds = TSDataset(df, freq="1d")
+
+    return tsds
+
+
+@pytest.mark.parametrize("model", [DeadlineMovingAverageModel])
+def test_simple_model_forecaster_run(simple_df, model):
+    sma_model = model(window=1)
```
julia-shenshina, line 33: not `sma_model`

julia-shenshina: can you just add `DeadlineMovingAverageModel` to test https://github.com/tinkoff-ai/etna/blob/master/tests/test_models/test_simple_models.py#L13 ?

tests/test_models/test_deadline_ma.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
```
julia-shenshina, line 1: add a test for model usage inside of Pipeline.backtest please

etna/models/__init__.py
```diff
@@ -8,6 +8,8 @@
 from etna.models.catboost import CatBoostModelPerSegment
 from etna.models.catboost import CatBoostMultiSegmentModel
 from etna.models.catboost import CatBoostPerSegmentModel
+from etna.models.deadline_ma import DeadlineMovingAverageModel
+from etna.models.deadline_ma import _DeadlineMovingAverageModel
```
julia-shenshina, line 12: actually you don't need to import "private" class `_DeadlineMovingAverageModel` here, only `DeadlineMovingAverageModel`

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+from typing import List
+
+import numpy as np
+import pandas as pd
+
+from etna.models.base import PerSegmentModel
+
+
+class _DeadlineMovingAverageModel:
+    """Seasonal moving average model that uses exact previous dates to predict."""
```
julia-shenshina, line 12: not `Seasonal moving average model`

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,148 @@
...
+        ----------
+        window: int
+            Number of values taken for forecast for each point.
+        seasonality: str
+            Monthly or annual seasonality.
+        """
+        self.name = "target"
+        self.window = window
+        if seasonality not in ["month", "year"]:
+            raise ValueError("Incorrect type of seasonality")
```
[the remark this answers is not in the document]
julia-shenshina: and maybe we should create a `class SeasonalityMode(str, Enum)` to work with seasonalities here like https://github.com/tinkoff-ai/etna/blob/master/etna/metrics/base.py#L14 ?
what do you think about it?

julia-shenshina (changes requested):

julia-shenshina:
update CHANGELOG please

etna/models/deadline_ma.py
```diff
@@ -0,0 +1,164 @@
...
+            for w in range(1, self.window + 1):
+                if self.seasonality == SeasonalityMode.month:
+                    prev_date = res.index[i] - pd.DateOffset(months=w)
+
+                elif self.seasonality == SeasonalityMode.year:
+                    prev_date = res.index[i] - pd.DateOffset(years=w)
+                if prev_date <= self.timestamps.iloc[-1]:
+                    res.loc[index[i]] += self.series.loc[self.timestamps == prev_date].values
+                else:
+                    res.loc[index[i]] += res.loc[prev_date].values
```
julia-shenshina, line 122: could you please add a test for this "else" behavior?

smth like
```
pipeline = Pipeline(model=DeadlineMovingAverageModel(window=5, seasonality="month"), transforms=[], horizon=180)
```
and check that everything works

julia-shenshina (dismissed):

julia-shenshina (approved):

PR #892, opened 2022-08-30
update holiday transform

DBcreator:

### DIFF
--- a/etna/transforms/timestamp/holiday.py
+++ b/etna/transforms/timestamp/holiday.py
@@ -1,4 +1,5 @@
 import datetime
+from typing import List
 from typing import Optional
 
 import holidays
@@ -6,8 +7,8 @@ import numpy as np
 import pandas as pd
 
 from etna.transforms.base import FutureMixin
-from etna.transforms.base import IrreversibleTransform, IrreversiblePerSegmentWrapper
-from typing import List
+from etna.transforms.base import IrreversibleTransform
+
 
 class HolidayTransform(IrreversibleTransform, FutureMixin):
     """HolidayTransform generates series that indicates holidays in given dataframe."""
@@ -71,6 +72,7 @@ class HolidayTransform(IrreversibleTransform, FutureMixin):
         df = df.join(encoded_df)
         df = df.sort_index(axis=1)
         return df
+
     def get_regressors_info(self) -> List[str]:
         """Return the list with regressors created by the transform.
         Returns
@@ -78,4 +80,4 @@ class HolidayTransform(IrreversibleTransform, FutureMixin):
         :
             List with regressors created by the transform.
         """
-        return [self.out_column]
+        return [self.out_column]  # type: ignore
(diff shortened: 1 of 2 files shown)


etna/transforms/timestamp/holiday.py
```diff
@@ -23,12 +24,13 @@ def __init__(self, iso_code: str = "RUS", out_column: Optional[str] = None):
         out_column:
             name of added column. Use ``self.__repr__()`` if not given.
         """
+        super().__init__(required_features="all")
```
alex-hse-repository, line 27: This transform actually does not require any features, so there should be ["target"] just as placeholder

etna/transforms/timestamp/holiday.py
```diff
@@ -70,3 +72,12 @@ def transform(self, df: pd.DataFrame) -> pd.DataFrame:
...
         return df
+
+    def get_regressors_info(self) -> List[str]:
+        """Return the list with regressors created by the transform.
+        Returns
+        -------
+        :
+            List with regressors created by the transform.
+        """
+        return [self.out_column]  # type: ignore
```
alex-hse-repository, line 83: To fix the problem with typing you should remove this line `self.out_column = out_column` from the constructor and fix the definition on the next line. This happens because mypy think that out_column might be None, but in fact it is not.

tests/test_transforms/test_timestamp/test_holiday_transform.py
```diff
@@ -84,33 +84,37 @@ def two_segments_simple_df_min(simple_constant_df_min: pd.DataFrame):
...
+    holiday = HolidayTransform(out_column="holiday")
+    new = holiday.fit_transform(simple_ts_with_regressors)
+    len_holiday = len([cols for cols in new.columns if cols[1] == "holiday"])
+    assert len_holiday == len(np.unique(new.columns.get_level_values("segment")))
 
 
 def test_interface_two_segments_daily(two_segments_simple_df_daily: pd.DataFrame):
     holidays_finder = HolidayTransform(out_column="regressor_holidays")
-    df = holidays_finder.fit_transform(two_segments_simple_df_daily)
+    ts = TSDataset(two_segments_simple_df_daily, freq="D")
```
alex-hse-repository, line 95: You should fix the fixture itself, now it should be `two_segments_simple_ts_daily `

tests/test_transforms/test_timestamp/test_holiday_transform.py
```diff
@@ -84,33 +84,37 @@ def two_segments_simple_df_min(simple_constant_df_min: pd.DataFrame):
...
+    df = holidays_finder.fit_transform(ts)
```
alex-hse-repository, line 96: This is now not df, this is also ts. You should use `to_pandas` method to get the df

tests/test_transforms/test_timestamp/test_holiday_transform.py
alex-hse-repository: Fix all the fixtures used in these tests

tests/test_transforms/test_timestamp/test_holiday_transform.py
alex-hse-repository: When you get the df from TSDataset the code below should work without changes

alex-hse-repository (changes requested):

alex-hse-repository (approved):
👍

PR #984, opened 2022-10-13
New slice behaviour of `TSDataset`

### LINKED ISSUE #968: Make `TSDataset` slice return `TSDataset`
### 🚀 Feature Request

Now the slice form TSDataset should also be TSDataset

### Proposal

1. Change current behaviour of `__getitem__`, `head`, `tail` methods -- now it should return the corresponding TSDataset
2. Method `__getitem__` should work using the following rules:
- If 1 item -- [timestamp]
- If 2 items -- [timestamp, feature]
- If 3 items -- [timestamp, segment, feature]
- timestamp = int, str, slice[int], slice[str]
- segment/feature = str, list[str] or ":"
4. Remove methods `loc`, `tsdataset_idx_slice`
5. Fix all the places, where the slice of dataset is used including notebooks
7. Improve the existing tests

### Test cases

1. Method `head(10)` works the same as `ts[-10:]`
2. Method `tail(10)` works the same as `ts[:10]`
3. Test all the possible cases of `__getitem__`

### Additional context

_No response_

alex-hse-repository:

## Closing issues
closes #968

### DIFF
--- a/etna/datasets/tsdataset.py
+++ b/etna/datasets/tsdataset.py
@@ -179,24 +179,28 @@ class TSDataset:
                 item = [item]
             return item
 
-        df_exog = self.df_exog.copy()
+        df_exog = self.df_exog
         if isinstance(item, slice) or isinstance(item, int) or isinstance(item, str):
             time_idx = to_time_index(item)
             df = self.df.loc[self.idx[time_idx]]
+            raw_df = self.raw_df.loc[self.idx[time_idx]]
         elif isinstance(item, tuple):
             time_idx = to_time_index(item[0])
             if len(item) == 2:
                 df = self.df.loc[self.idx[time_idx], self.idx[:, item[1]]]
+                raw_df = self.raw_df.loc[self.idx[time_idx]]
             elif len(item) == 3:
-                segments = [item[1]] if isinstance(item[2], str) else item[1]
-                features = [item[2]] if isinstance(item[2], str) else item[2]
+                segments = (item[1]) if isinstance(item[1], str) else tuple(item[1])
+                features = (item[2]) if isinstance(item[2], str) else tuple(item[2])
                 df = self.df.loc[self.idx[time_idx], self.idx[segments, features]]
-                df_exog = self.df_exog.loc[self.idx[:], self.idx[segments, :]]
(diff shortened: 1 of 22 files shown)


etna/datasets/tsdataset.py
```diff
@@ -165,18 +166,60 @@ def __repr__(self):
...
+        ts.known_future = deepcopy(self.known_future)
+        ts._regressors = deepcopy(self.regressors)
+        return ts
+
+    def __eq__(self, other):
+        return (
+            sorted(self.known_future) == sorted(other.known_future)
+            and sorted(self.regressors) == sorted(other.regressors)
+            and (self.df == other.df).all().all()
+            and (self.df_exog == other.df_exog).all().all()
```
Ama16, line 221: If df_exog == None it will be raised error on .all()

etna/datasets/tsdataset.py
```diff
@@ -165,18 +166,60 @@ def __repr__(self):
...
+                df = self.df.loc[self.idx[time_idx], self.idx[segments, features]]
+                df_exog = None if df_exog is None else df_exog.loc[self.idx[:], self.idx[segments, :]]
+                raw_df = self.raw_df.loc[self.idx[time_idx], self.idx[segments, :]]
+            else:
+                raise ValueError("Unsupported indexing format!")
         else:
-            df = self.df.loc[self.idx[item[0]], self.idx[item[1], item[2]]]
+            raise ValueError("Unsupported indexing format!")
+
         first_valid_idx = df.first_valid_index()
```
martins0n, line 207: It seems this is not expected. 
We don't need it, because None at the begining make sense as for upstream TSDataset

etna/datasets/tsdataset.py
alex-hse-repository: I removed this magic method at all, I am not sure we really need it

etna/datasets/tsdataset.py
alex-hse-repository: We do't need None at the begining, in the old version we also cut them in getitem

martins0n: Old version is not the target. It was developed in a hurry)

We cut them before, because this slices were used for persegment models mostly.

But it's important to note that there was and there is an idea that None at beginnig are about missed values of some process and we shouldn't just drop them without particular reason

PR #1181, opened 2023-03-23
Implement in-sample predictions in BATS/TBATS

brsnw250:


etna/models/tbats.py
```diff
@@ -65,7 +65,31 @@ def forecast(self, df: pd.DataFrame, prediction_interval: bool, quantiles: Itera
...
     def predict(self, df: pd.DataFrame, prediction_interval: bool, quantiles: Iterable[float]) -> pd.DataFrame:
-        raise NotImplementedError("Method predict isn't currently implemented!")
+        if self._fitted_model is None or self._freq is None:
+            raise ValueError("Model is not fitted! Fit the model before calling predict method!")
+
+        y_pred = pd.DataFrame()
+        y_pred["target"] = self._fitted_model.y_hat
+        y_pred["timestamp"] = pd.date_range(end=str(self._last_train_timestamp), freq=self._freq, periods=len(y_pred))
+
+        if len(set(y_pred["timestamp"]) & set(df["timestamp"])) == 0:
```
alex-hse-repository, line 75: I guess there should be `set(df["timestamp"]) <= set(y_pred["timestamp"])`, otherwise we can't predict all the timestamps in passed dataframe

alex-hse-repository: Can't we check it in the begining and just fill "target" in passed df?

etna/models/tbats.py
```diff
@@ -114,7 +138,15 @@ def predict_components(self, df: pd.DataFrame) -> pd.DataFrame:
         :
             dataframe with prediction components
         """
-        raise NotImplementedError("Prediction decomposition isn't currently implemented!")
+        if self._fitted_model is None or self._freq is None:
+            raise ValueError("Model is not fitted! Fit the model before estimating forecast components!")
+
+        self._check_components()
+
+        raw_components = self._decompose_predict()
```
alex-hse-repository, line 146: Here we should decompose predictions only on `df`, now you get the decomposition from the whole train dataset

etna/models/tbats.py
```diff
@@ -175,9 +217,33 @@ def _decompose_forecast(self, horizon: int) -> np.ndarray:
...
 
         if model.params.components.use_box_cox:
-            transformed_pred = np.sum(raw_components, axis=1)
-            pred = model._inv_boxcox(transformed_pred)
-            raw_components = raw_components * pred[..., np.newaxis] / transformed_pred[..., np.newaxis]
+            raw_components = self._rescale_components(raw_components)
+
+        return raw_components
+
+    def _decompose_predict(self) -> np.ndarray:
```
alex-hse-repository, line 224: In in-sample predictions we also have `residuals` component as I understand

tests/test_models/test_tbats.py
```diff
@@ -203,6 +232,23 @@ def test_decompose_forecast_output_format(small_periodic_ts, estimator):
...
+    ),
+)
+def test_decompose_predict_output_format(small_periodic_ts, estimator):
+    model = estimator()
+    model.fit(small_periodic_ts)
+
+    target = small_periodic_ts[:, "segment_1", "target"].values
+    components = model._models["segment_1"]._decompose_predict()
+    assert isinstance(components, np.ndarray)
+    assert components.shape[0] == target.shape[0]
```
alex-hse-repository, line 249: This is true only if we decompose on full train, try to decompose on the part of the train

tests/test_models/test_tbats.py
```diff
@@ -296,17 +376,24 @@ def test_arma_component_not_fitted(small_periodic_ts, estimator):
         TBATSModel,
     ),
 )
-def test_arma_w_seasonal_components_not_fitted(small_periodic_ts, estimator):
+@pytest.mark.parametrize("method,use_future", (("predict_components", False), ("forecast_components", True)))
+def test_arma_w_seasonal_components_not_fitted(small_periodic_ts, estimator, method, use_future):
```
alex-hse-repository, line 380: w can stands for "with" and "without" in the same time

tests/test_models/test_tbats.py
```diff
@@ -258,16 +324,23 @@ def test_components_names(periodic_ts, estimator, params, components_names):
...
+@pytest.mark.parametrize("method,use_future", (("predict_components", False), ("forecast_components", True)))
+def test_seasonal_components_not_fitted(small_periodic_ts, estimator, method, use_future):
     model = estimator(seasonal_periods=[7, 14], use_arma_errors=False)
     model.fit(small_periodic_ts)
 
-    future = small_periodic_ts.make_future(3).to_pandas(flatten=True)
     segment_model = model._models["segment_1"]
     segment_model._fitted_model.params.components.seasonal_periods = []
 
+    pred_ts = small_periodic_ts
```
alex-hse-repository, line 335: This pattern occurs in several tests, do we really need it? I guess it makes tests overcomplicated. Better to test all the methods on appropriate fixtures in long format and use adapters instead of models

alex-hse-repository: We can create separate fixtures for train and test, do't generate them inside the tests with make_future

alex-hse-repository (changes requested):

etna/models/tbats.py
brsnw250: No, prediction formed only from components of state vector. Residuals are `y - y_hat`. ([here](https://github.com/intive-DataScience/tbats/blob/1f2e0b5e769250c8ec0604fd75ef08ebbe251d37/tbats/abstract/Model.py#L236))

.github/workflows/test.yml
```diff
@@ -148,6 +148,44 @@ jobs:
       - name: Upload coverage
         uses: codecov/codecov-action@v2
 
+  long-3-test:
```
alex-hse-repository, line 151: Not the best practice, we should use template for this purpose. We will do it in the separate task I guess

tests/test_models/test_tbats.py
```diff
@@ -144,22 +149,51 @@ def test_format(model, new_format_df):
 
 @pytest.mark.long_2
 @pytest.mark.parametrize("model", [TBATSModel(), BATSModel()])
-def test_dummy(model, sinusoid_ts):
+def test_predict_format(model, new_format_df):
```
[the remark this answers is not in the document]
alex-hse-repository: Why did you just remove it but not parametrize `test_forecast_format`?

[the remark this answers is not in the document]
brsnw250: This test was just to test behaviour with transform. We have tests for general `predict` behaviour in inference tests.

tests/test_models/test_tbats.py
brsnw250: This is always true due to specific of `_decompose_predict`. I've added tests for part of train.

tests/test_models/test_tbats.py
brsnw250: This pattern is a result of parametrization with `forecast_decompose` and `predict_decompose`. Now tests updated and it looks a bit different.

alex-hse-repository (changes requested):

.github/workflows/test.yml
brsnw250: ok. will remove this then

tests/test_models/test_tbats.py
```diff
@@ -269,6 +269,18 @@ def test_named_components_output_format(small_periodic_ts, estimator):
     assert len(components) == horizon
 
 
+@pytest.mark.parametrize(
+    "train_slice,decompose_slice", ((slice(5, 20), slice(None, 20)), (slice(5, 10), slice(10, 20)))
```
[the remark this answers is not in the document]
brsnw250: Here we have only 2 cases: when train and decompose dataframes has intersection and without it.

alex-hse-repository (approved):

